PR #1, opened 2019-07-08
Please review Day5!

Shazmaan:


Problem2.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <map>
+
+using namespace std;
+
+class Solution{
+private:
+    map<char,char> hashmap;
+    set<char> hashset;
+public:
+    bool isIsomorphic(string s, string t) {
```
super30admin: Shazmaan, You should comment your approach along with the code.

Shazmaan: I will start doing that starting now. I apologize for the inconvenience!

Shazmaan: added comments, please view.

PR #6, opened 2019-07-09
completed group anagram isomorphicString and Patternmatch- please review

PremalSamale:


GroupAnagram.java
```diff
@@ -0,0 +1,38 @@
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.HashMap;
+import java.util.Map;
+
+//TIME COMPLEXITY O(nklogk) I did not understand
```
ashmeetdeol19: n : for no. of elements
klogk : for sorting part (k is length of string) (java uses quickSort for primitive data types and mergeSort for objects)

For each element we do sorting So it comes out to be n*klogk

Is it clear now ?

GroupAnagram.java
```diff
@@ -0,0 +1,38 @@
...
+//space complexity o(nk) ) I did not understand
```
ashmeetdeol19: k: average length of strings
n: no. of strings

So total space required = n*k

Is it clear now?

IsomorphicString.java
```diff
@@ -0,0 +1,75 @@
...
+
+//reference https://www.geeksforgeeks.org/check-if-two-given-strings-are-isomorphic-to-each-other/
+
+//time complexity o(n)
+//space complexity o(n)
+import java.io.*; 
+import java.util.*; 
+class IsomorphicString 
+{ 
+	static int size = 256;
```
ashmeetdeol19: Can you do it with size 26 ?

PR #7, opened 2019-07-09
Completed Problem 1 and 2 -- Please review. both are O(n)

lavanganji:


AnagramGrouping.java
```diff
@@ -0,0 +1,52 @@
+
+// Time Complexity is O(n)
```
ashmeetdeol19: Kindly think about this once again? Think about Array.sort() you are using.

isomorphic.java
```diff
@@ -0,0 +1,32 @@
+public class isomorphic {
+
+	public boolean isIsomorphic(String s, String t) {
+        int[] a1 = new int[256];
```
ashmeetdeol19: Try it with int[26] if its possible.

PR #9, opened 2019-07-09
Completed Problem 1, please review

Ameya145:


Problem1
```diff
@@ -0,0 +1,25 @@
+//Time complexity: O(nlgn)
```
ashmeetdeol19: Can you explain how the complexity is nlogn and also mention space complexity for the same.

Problem 2
```diff
@@ -0,0 +1,21 @@
+//Time complexity : O(n)
+//Space complexity : O(n)
+
+class Solution {
+    public boolean isIsomorphic(String s, String t) {
+        int n = s.length(), m = t.length();
+        if(n != m) return false;
+        int[] map = new int[256];
```
ashmeetdeol19: Think, If you can do this with size=26.

Problem 3
```diff
@@ -0,0 +1,28 @@
...
+            } else if(check !=null && s!= null) {
+                if(!words[i].equals(s)) return false;
+                if(!(ch==check)) return false;
+            } else {
+                return false;
+            }
+        }
+        return true;
+    }
+}
```
ashmeetdeol19: Well Done!!

PR #10, opened 2019-07-09
Please review Problem 1 and Problem 2 (Completed)

AkhilKrishnamurthy:


problem2.java
```diff
@@ -0,0 +1,52 @@
...
+
+    }
+
+
+    public static void main(String[] args) {
+        String s = "egg";
+        String t = "ada";
+        System.out.println(isomorphic(s,t));
+    }
+}
```
ashmeetdeol19: This is the new solution that you provided us but unfortunately its not the optimal one.
Keep thinking and keep coding 👍

PR #19, opened 2019-08-06
Hashing 1 Completed. Please Review

priyansh22:


Problem3.java
```diff
@@ -0,0 +1,32 @@
...
+    if(s.length() != str.length()) {
+        return false;
+    }
+
+    HashMap<Character, String> h1 = new HashMap<Character, String>();
+    HashMap<Character, String> h2 = new HashMap<Character, String>();
+String[] result = str.split(" ",2);
+    for(int i = 0; i < s.length(); i++) {
+        if(h1.get(s.charAt(i))!=null) {
+        if(h1.get(s.charAt(i))!=(result[i])) return false;
```
kaurramanpreet: at this place if you are using != or == it will be comparing the references to those values not the values itseld
so as the values are placed at different locations so it wont be comparing the values instead it would be comparing the references and hence will give the answer as false

Problem3.java
```diff
@@ -0,0 +1,32 @@
...
+Runs on LeetCode : NO 
+
+class Solution {
+    public boolean wordPattern(String s, String str) {
+    if(s.length() != str.length()) {
+        return false;
+    }
+
+    HashMap<Character, String> h1 = new HashMap<Character, String>();
+    HashMap<Character, String> h2 = new HashMap<Character, String>();
```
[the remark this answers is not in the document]
priyansh22: Yes the declaration should be changed. Thank you

Problem3.java
priyansh22: Yes understood. Will change it. 
Thank You

PR #20, opened 2019-08-26
Completed" Kindly review the anagram problem"

DeekanshaTandon:


isomorphic.java
```diff
@@ -0,0 +1,33 @@
...
+            return false;
+
+        HashMap<Character,Integer > map1 = new HashMap<>();
+        HashMap<Character,Integer > map2 = new HashMap<>();
+
+        for(int i =0;i<s.length();i++){
+            Character s1= s.charAt(i);
+            Character t1 =t.charAt(i);
+
+            if(!map1.getOrDefault(s1,-1).equals(map2.getOrDefault(t1,-1))){
```
kaurramanpreet: are you comparing the indices and not key values?
should you be having integer as values?

kaurramanpreet: but if you are accessing using maps that means you are accessing the index 
ping me if you can have a zoom call
we will discuss your approach as well

PR #22, opened 2019-08-27
Problem 1 and 2 Done

pradhan1234:
cc @kaurramanpreet


Isomorphic.java
```diff
@@ -18,10 +18,20 @@
  * Runtime: 11 ms, faster than 42.82% of Java online submissions for Isomorphic Strings.
  * Memory Usage: 36.4 MB, less than 100.00% of Java online submissions for Isomorphic Strings.
  * 
+ * 
  * Another approach with same tc but better performance would be to use a
  * character array to preserve mappings. Overhead of computing hashes is reduced.
```
kaurramanpreet: good to see you are able to think about the question from different perspectives as well!!

PR #36, opened 2019-10-15
word pattern - problem

vramanathan:
sorry i misunderstood today's assignments. I just have only one out of 3 done


vramanathan:
added problems 1 and 2

PR #44, opened 2019-10-17
hashing-1

priyanshi-1234:


isomorphicstrings.cpp
```diff
@@ -0,0 +1,61 @@
...
+            else
+            {
+                if(mapp[key]==key1)
+                {
+                    return true ;
+
+                }
+                else
+                     return false;
+            }
```
sehaj-grover: You can't return at the first check 
Just check the false cases not the true case
'aab' and 'abb' you return true at the first index and don't check the rest

wordpattern.cpp
```diff
@@ -0,0 +1,66 @@
...
+        string temp="";
+         vector<string> v;
+        int count=0;
+      for(int i=0;i<str.length();i++)
+    {
+          if(str[i]!=' ')
+         {
+
+              temp+=str[i];
+              count++;
```
sehaj-grover: count should not be increment in the if but in the else

wordpattern.cpp
```diff
@@ -0,0 +1,66 @@
...
+         }
+     else
+     {
+     v.push_back(temp);
+     temp="";
+     }
+
+    }
+
```
sehaj-grover: plus you are adding only the n-1 words into the vector the last word is missed because of your if condition so you have to add the last word outside the for loop

wordpattern.cpp
```diff
@@ -0,0 +1,66 @@
...
+            if(hash1.count(pattern[i])!=0)
+            {
+               if( hash1[pattern[i]]!=v[i])
+               {
+                   return false;
+               }
+                else
+                    return true;
+            }
```
sehaj-grover: check only the false cases

PR #45, opened 2019-10-20
Completed Hashing 1

nshivapr:
Please review

### DIFF
--- a/isIsomorphic.java
+++ b/isIsomorphic.java
@@ -1,0 +1,34 @@
+// 205. Isomorphic Strings
+// Time Complexity : O(n)
+// Space Complexity : O(1)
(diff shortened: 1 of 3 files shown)


isIsomorphic.java
```diff
@@ -0,0 +1,34 @@
...
+    public boolean isIsomorphic(String s, String t) {
+        HashMap<Character, Character> sHash = new HashMap<>();
+        HashMap<Character, Character> tHash = new HashMap<>();
+
+        if(s.length() != t.length())
+            return false;
+
+        for(int i=0; i<s.length(); i++){
+            char a = s.charAt(i);
+            char b = s.charAt(i);
```
sehaj-grover: you are taking both a and b from string s 
b should be t.charAt(i)

nshivapr:
Got it

PR #51, opened 2019-11-26
Pushing Hashing 1 changes

dasgprat:
Please review the PR and also let me know in case of any issues or queries

### DIFF
--- a/Problem1.java
+++ b/Problem1.java
@@ -1,0 +1,53 @@
+
+/* The idea is to use index to decide which part of the array there will be missing element and then traverse accordingly*/
+
+class solution{
+    
+    public static void main(String [] args) {
+	//code
+	 int[] nums = {0,1,2,3,5,6};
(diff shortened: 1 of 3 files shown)


Problem1.java
```diff
@@ -0,0 +1,53 @@
...
+            if(a[mid] == mid)
+                return search(a,mid+1, r);
+            else
+                return search(a,l,mid - 1);
+        
+        }
+        return -1;
+
+    
+}
```
neha79091: This logic will not work if the list doesn't start with 0

dasgprat:
Yeah It seems I committed wrong code in problem 1 .
Added the anagram code and Binary search code is already there in competitive coding

Problem1.java
dasgprat: In my mock interview I did the version where list starts with 0 from Leetcode.
Can update the code (if required)

PR #80, opened 2020-01-06
Please review hashing

BhaveshTadikonda:


super30admin:
The space in isomorphic strings is O(1) and not O(n). Kindly try word pattern and other solution of group anagrams as well

Word pattern.java
```diff
@@ -0,0 +1,35 @@
+// Time Complexity - 0(n)
+// space Complexity - 0(n)
+import java.io.*;
```
neha79091: Space complexity is constant as there will be only 26 alphabets at maximum

PR #111, opened 2020-03-18
Completed Hashing 1

arcsoftech:


Isomorphic.java
```diff
@@ -0,0 +1,41 @@
+import java.util.HashMap;
+
+// Time Complexity :O(n)
+// Space Complexity :O(n) + O(n)=> O(n) -- For two Hashmap.
```
devashishnyati: The space complexity here will be constant because at max you can have 26 keys.

GroupAnagram.java
```diff
@@ -0,0 +1,29 @@
+// Time Complexity :O(n)
```
devashishnyati: You did not add the time complexity of signature function here.

devashishnyati:
Good job Arihant!
Please review the time and space complexities.

Isomorphic.java
arcsoftech: Yeah! you are right it should be O(1).

GroupAnagram.java
arcsoftech: Here space should be O(n) because in worst case it my be possible that each word does not belong to any group. In that case it will be O(n).  Time will be O(nk) where n is the number of strings and k will be maximum length of string . Correct ?

devashishnyati: Yes Arihant

PR #112, opened 2020-03-18
Please Review Problem-1

AMAbhi:


problem1.py
```diff
@@ -0,0 +1,28 @@
+# Time Complexity : O(1)
```
devashishnyati: The time complexity is not O(1)

devashishnyati:
Nice try Abhishek!
Please submit all the questions with correct time and space complexities.

PR #114, opened 2020-03-19
Please review - Hashing 1

megrao:
I was able to 1 & 3 exercises. For 2, I have written a pseudocode but my code is way too similar to exercise 1 and fails in Leetcode.

### DIFF
--- a/Hashing1_Exercise1.py
+++ b/Hashing1_Exercise1.py
@@ -1,0 +1,29 @@
+# Time Complexity : O(log(n)) ? #Not sure
+# Space Complexity : O(n)
+# Did this code successfully run on Leetcode : Yes
+# Challenges: Initially, I wasn't sure how to proceed.
+
+Comments: I am creating a hashmap with sorted string as its key and value as new strings.
+
+class Solution(object):
+    def groupAnagrams(self, strings):
+        dict = {}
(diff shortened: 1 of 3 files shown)


Hashing1_Exercise1.py
```diff
@@ -0,0 +1,29 @@
+# Time Complexity : O(log(n)) ? #Not sure
```
devashishnyati: You are missing the time complexity of sorting the string.
Also, it won't be O(logn). You will see logn complexity when at each step you are eliminating the half of the choices. Here you are iterating through all the strings. So it will be O(n klogk) where klogk is the time complexity to sort the strings.

Hashing1_Exercise1.py
```diff
@@ -0,0 +1,29 @@
...
+        dict = {}
+        result = []
+        for str in strings:
+            sorted = ''.join(sorted(str))
+            if sorted in dict:
+                dict[sorted].append(str)
+            else:
+                dict[sorted] = [str]
+        
+        for values in dict.values():
```
devashishnyati: You can directly return dict.values() also. Try it out

devashishnyati:
Good job Megha!
Please submit the other problem also.

PR #128, opened 2020-04-25
Completed Problems: Please Review

sbsreedh:


anagram.py
```diff
@@ -0,0 +1,27 @@
+
+# from collections import defaultdict
+# TC: O(n)
+# SC: O(n)
+# Successfully ran on Leetcode: Yes
+# Difficulties: I am doubtful about TC here. Is it O(n*len(item in array))??
```
iivaishnavii, line 6: Yes,it is O(n*k) -where n is the size of the array and k is the length of each string

iivaishnavii:
Great job, Sonali! Take a look at the inline comments as well.

PR #129, opened 2020-04-25
Completed The Design for First and Second Question

JacobStephen9999:


iivaishnavii:
Great job, Jacob! Do complete the 3rd question as well.

PR #131, opened 2020-04-25
Hashing 1 done

navnitss:


Sample.java
```diff
@@ -0,0 +1,88 @@
+// Time Complexity :O(n)
```
iivaishnavii, line 1: For each word in the array, you are again sorting it.. Update time complexity please.

Sample.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+---------------------------------------------------------------------------------
+
+// Time Complexity :O(n)
+// Space Complexity :O(1)
+// Your code here along with comments explaining your approach	
+public class Solution {
+    public boolean isIsomorphic(String s, String t) {
+        if(s == null || s.length() <= 1) return true;
```
iivaishnavii, line 31: Handle the case of both the lengths being the same.

Sample.java
```diff
@@ -0,0 +1,88 @@
...
+        HashMap<Character, Character> map = new HashMap<Character, Character>();
+        for(int i = 0 ; i< s.length(); i++){
```
iivaishnavii, line 33: what if both strings don't have same length?

iivaishnavii:
Good job,Navnit!

PR #143, opened 2020-04-28
Problem12 - Please review

KGurpreet:


Problem12
```diff
@@ -0,0 +1,31 @@
+// Time Complexity : O(log n)
```
super30admin, line 1: The time is incorrect. its O(n)

super30admin: Want to know your doubt....Why do you believe its log (n)?

super30admin:
Good. Please view inline comment

PR #146, opened 2020-04-29
Completed Anagrams and Isomorphic strings - Please Review

karthikmk:


word_pattern.cc
```diff
@@ -0,0 +1,36 @@
+// Time Complexity : O(n) - n is length of each string. It is O(1) for lookup and insertion.
+// Space Complexity : O(1) - We use 2 maps of constant size always no matter what the length of the word string is (controlled by char map)
```
sowmyadvn, line 2: Map<char, string> is always of constant size, but Map<string, char> size varies. So, space complexity would be O(N) in this case

sowmyadvn:
Solutions are correct. Please check comments on space complexity for word pattern problem

sowmyadvn:
Solutions are correct. Please check comments on space complexity for word pattern problem

PR #148, opened 2020-04-29
Problem 1 and 2 done.Please review

shriyavanvari:


Problem1.py
```diff
@@ -0,0 +1,16 @@
+#Time Complexity: O(Klog N) where k-no of string given in the list, log n for sorting strings of length n
```
super30admin, line 1: This is incorrect. Please watch video again

super30admin: its knlogn by the way you have defined k and n

Problem2.py
```diff
@@ -0,0 +1,21 @@
+#Space Complexity: O(n)  where n is  length of str: For 2 dictionaries
+#Time Complexity: O(n)- where n is  length of str
```
super30admin, line 2: O(1). Can you ping me on slack. I am not sure you are understanding things in class...Both problems have wrong time and space and is a red flag for interviews

super30admin:
Kindly review inline comments and work hard.

PR #159, opened 2020-05-28
Create GroupAnagrams_Isomorphic_WordPattern.py

tramatejaswini:


GroupAnagrams_Isomorphic_WordPattern.py
```diff
@@ -0,0 +1,84 @@
...
+
+# Time Complexity : O(nk logk) n is the number of strings and k is the max characters of strings
+# Space Complexity : O(nk)
+# Did this code successfully run on Leetcode : Yes
+# Any problem you faced while coding this : No
+# 
+# Used dictionary to store the values of the tuple.
+
+class Solution:
+    def groupAnagrams(self, strs: List[str]) -> List[List[str]]:
```
AbhiniveshP, line 14: Try using prime products as keys which will make your solution more efficient !!!

AbhiniveshP:
Good job Rama Tejaswini !!!
Try avoiding small libraries like defaultdict() and zip() which can be implemented quickly. Instead, try implementing on your own which makes your code even better.

PR #161, opened 2020-05-28
Hashing - 1 Initial Commit - Review my submission

Shoban-23:


Problem-2.java
```diff
@@ -0,0 +1,19 @@
+class Solution {
```
chiragshahi, line 1: follow start template

chiragshahi:
Excellent work Shoban!

PR #165, opened 2020-05-29
PA_11.py

trivedipratul:


PA_11.py
```diff
@@ -0,0 +1,15 @@
+#LC 49 - Group Anagrams
+#Time Complexity - n*(slogs) s - len(string) n - len(strs)
+#Space Complexity - O(2n)
+from collections import defaultdict
```
Lohith-Aswathappa, line 4: Great work limiting the solution to very few lines.

Lohith-Aswathappa:
Good job, Pratul!
Try the other 2 problems as well before the class

PR #168, opened 2020-05-29
Completed Hashing-1. Kindly review and share feedback.

Sumesh-H:


Anagrams.java
```diff
@@ -0,0 +1,29 @@
+// Time Complexity : O(N*K*log K), where N is the number of token, K is length of longest token
```
chiragshahi, line 1: try without sorting as well

chiragshahi:
Good work! See inline comment

Anagrams.java
Sumesh-H: Sure, will do

PR #176, opened 2020-05-29
Abhijeet - Hashing1

AbhijeetDhillon:


groupAnagrams.java
```diff
@@ -0,0 +1,43 @@
+// Time Complexity : O(nm) n is the array size and m is the character length in a word for this operation "Arrays.sort(word)"
```
chiragshahi, line 1: sorting is done internally with a complexity of nlgn
So the total complexity here would be nmlgm where m is the word length and n is the number of words
but we can do this problem more efficiently in mn

isIsomorphic.java
```diff
@@ -0,0 +1,40 @@
+// Time Complexity : O(nm) n is the array size and m is the character length in a word for this operation "hashmap.containsKey" 
+// Space Complexity : O(n) since the number elements in the array are used multiple times and hence don't have a fixed usage count and also because of auxilary data structures like hashmap and list being used.
```
chiragshahi, line 2: O(1) since the map can only have 26 KV pairs

wordPattern.java
```diff
@@ -0,0 +1,42 @@
+// Time Complexity : O(nm) n is the array size and m is the character length in a word for this operation "hashmap.containsKey" 
+// Space Complexity : O(n) since the number elements in the array are used multiple times and hence don't have a fixed usage count and also because of auxilary data structures like hashmap and list being used.
```
chiragshahi, line 2: O(1)

chiragshahi:
Well done! See inline comment

PR #183, opened 2020-06-01
Please review it

nareshkumar-ss:


Anagrams.java
```diff
@@ -1,3 +1,10 @@
+//I guess this solution takes O(n) time and O(mn) space where m is the number of characters in each word
```
AbhiniveshP, line 1: if you are sorting the chars of a word, it would be O(m * logm * n)

AbhiniveshP:
Good job !!!
Please try using prime products as keys in Anagrams question which would be more optimal.
Also, please include a 3-step explanation of your approach at the top

PR #186, opened 2020-06-01
Completed Hashing-1: Please review

Akashbir:


GroupAnagrams.java
```diff
@@ -0,0 +1,31 @@
+// Time Complexity: O(m*nlogn) where m is the length of strs array and n is the length of each string in the array, nlogn for sorting each
+// string in the strs array.
+// Space Complexity: O(m): Length of strs array
```
chiragshahi, line 3: The input parameter of the function does not contribute to our space complexity that we denote here.
It depends on the auxiliary DS that we use for solving, in this case a hash map

IsomorphicStrings.java
```diff
@@ -0,0 +1,32 @@
+// Time Complexity: O(n): n is the length of string s and t.
+// Space Complexity: O(n): n is the length of string s or t.
```
chiragshahi, line 2: Same comment as previous question for space complexity.
Here again the hash map contributes to the space complexity which is O(n) but since n can be at most 26, the space complexity here is O(1)

IsomorphicStrings.java
```diff
@@ -0,0 +1,32 @@
+// Time Complexity: O(n): n is the length of string s and t.
```
chiragshahi, line 1: what if s and t are different in length?

WordPattern.java
```diff
@@ -0,0 +1,34 @@
+// Time Complexity: O(n): n is the length of string str
+// Space Complexity: O(n): n is the length of string str
```
chiragshahi, line 2: Same comment as previous ones

chiragshahi:
Good work! Please see inline comments to better understand time space complexities.

GroupAnagrams.java
```diff
@@ -0,0 +1,31 @@
+// Time Complexity: O(m*nlogn) where m is the length of strs array and n is the length of each string in the array, nlogn for sorting each
```
chiragshahi, line 1: Try this in m*n as well

PR #187, opened 2020-06-01
Please review

snigdharreddy:


AbhiniveshP:
Good Work !!!
Please try to use product of primes as keys in Anagrams question instead of sorting, as it would provide a more optimal solution.

PR #205, opened 2020-07-13
Solutions to 3 problem__please review!

khgupta:


Problem3.py
```diff
@@ -0,0 +1,39 @@
...
+        :type pattern: str
+        :type str: str
+        :rtype: bool
+        """
+        s = pattern
+        t = str.split()
+        #print(zip(s,t))
+        return len(set(zip(s,t)))==len(set(s))==len(set(t)) and len(s)==len(t)
+
+    #Time Complexity: O(n) where n is number of characters and number of words in string
```
samarthjj, line 22: O(n*log(n)), because of the sorting in the return statements

samarthjj:
Good work Khushboo, Please review the inline comments!

samarthjj:
Good work Khushboo, Please review the inline comments!

Problem3.py
khgupta: yes, I have mentioned that for my second approach. For the first solution, it should be O(n) as it's just getting lengths and comparing.

PR #217, opened 2020-07-14
Create Group Anagrams - Please review

firoz1905:
Time Complexity of approach 2 still O(n*2) because of the product for loop . Is the time complexity correct  as per my understanding ?

### DIFF
--- a/Group Anagrams
+++ b/Group Anagrams
@@ -1,0 +1,38 @@
+create a dic of similar anagrams, the keys are sorted strings of each anagrams
+create a list of the dic values using the keys
+return result
+
+Approach 1 : Using Hash Map & Sort
+Time Complexity: O(n(klogk)) (K- Avg length of the words] , n - no of words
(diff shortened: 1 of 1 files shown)


Group Anagrams
```diff
@@ -0,0 +1,38 @@
...
+        for i in strs:
+            s="".join(sorted(i))
+            lst[s]=lst.get(s, []) + [i]
+        result = [v for v in lst.values()]
+        return result
+        
+ 
+ Approach 2 : Create a list with  prime numbers for all the 26 characters of the string . Instead of sorting , we can use property of prime numbers
+ whoose product will be unique.
+ Time Complexity: O(n^2*k)=O(n^2) ######## Still i feel like it is not good !
```
Sumesh-H, line 21: O(nk) where n is the number of strings in input String array and k is the length of each string

Sumesh-H:
Good work, Firoz! Kindly check inline comments.

PR #218, opened 2020-07-14
Hashing-1 completed please review

Shilpi10:


PatternCheck.java
```diff
@@ -0,0 +1,38 @@
+//Time complexity = O(n)
```
samarthjj, line 1: The TC should be O(nk), because adding and looking up a String (Pattern Array) is not a constant time operation

samarthjj:
Nice work Shilpi, Please review the inline comments!

samarthjj:
Nice work Shilpi, Please review the inline comments!

PR #219, opened 2020-07-14
Completed all 3 problems. Please Review it !

sneharnair:


Problem-1_Group_Anagrams.py
```diff
@@ -0,0 +1,106 @@
...
+                
+        return anagram.values()
+            
+        
+
+# APPROACH - 2  - INTERMEDIATE SOLUTION 
+# Time Complexity : O(n k log k) where n is the total number of words and k is the length of the longest word. 
+# Space Complexity : O(nk) - keys of the dict are of length k (bounded by max length). It's O(n), what if each of the words in the list are unique anagrams. 
+# Did this code successfully run on Leetcode : Yes
+# Any problem you faced while coding this :tried to use set() as key but they are mutable in Python.
```
samarthjj, line 44: You could also have stored the sorted word as the key, was there an actual need for a tuple or set?

Problem-2_Isomorphic_string.py
```diff
@@ -0,0 +1,89 @@
+# APPROACH - 1: (MY APPROACH) DIDN'T WORK  
+# Time Complexity : O(max(s, t)) - s is the length of s ad t is the length of t 
+# Space Complexity : O(max(s, t))
```
samarthjj, line 3: Space should be O(1), because the number of uniques chars cannot exceed 26.

Problem-2_Isomorphic_string.py
```diff
@@ -0,0 +1,89 @@
...
+            else:
+                return False
+                
+  
+  
+  
+  
+# APPROACH - 2: CORRECT APPROACH  
+# Time Complexity : O(s) or O(t) - s is the length of s ad t is the length of t and to be a valid case to check, s == t.
+# Space Complexity : O(s) or O(t) due to hash_s and hash_t
```
samarthjj, line 49: Space should be O(1), because the number of uniques chars cannot exceed 26.

Problem-3_Word_Pattern.py
```diff
@@ -0,0 +1,88 @@
+# APPROACH 1: (MY APPROACH BEFORE CLASS)
+# Time Complexity : O(s) or O(p) where s is the length of str and p is the length of pattern. for valid cases s == p.
```
samarthjj, line 2: The TC should be O(nk), where n is the len(p) or len(s), where k is the length of the longest word in pattern.
Looking up and adding a string in a hashmap or set is not a constant time operation

Problem-3_Word_Pattern.py
```diff
@@ -0,0 +1,88 @@
...
+# Space Complexity : O(pk) as number of entries in dict is p  and k is the length of the longest string in str
```
samarthjj, line 3: The TC should be O(nk), where n is the len(p) or len(s), where k is the length of the longest word in str.
Looking up and adding a string in a hashmap or set is not a constant time operation.
The SC should be O(1), because the number of uniques chars in pattern cannot exceed 26.

Problem-3_Word_Pattern.py
```diff
@@ -0,0 +1,88 @@
...
+        return True
+
+
+
+
+
+
+# APPROACH 2: CLASS APPROACH (SIMILAR TO 1)
+# Time Complexity : O(s) or O(p) where s is the length of str and p is the length of pattern, for valid cases s == p.
+# Space Complexity : O(pk) as number of entries in hash_pattern is p, key is always of length 1 and k is the length of the longest string in str (values). Similarly, number of
```
samarthjj, line 51: The TC should be O(nk), where n is the len(p) or len(s), where k is the length of the longest word in str.
Looking up and adding a string in a hashmap or set is not a constant time operation.
The SC should be O(1), because the number of uniques chars in pattern cannot exceed 26.

samarthjj:
Good job Sneha, Please review the inline comments.
Awesome work on including all the different approaches for all the problems!

samarthjj:
Good job Sneha, Please review the inline comments.
Awesome work on including all the different approaches for all the problems!

PR #221, opened 2020-07-14
Please Review : Hashing-1

akashkshekarappa:


src/GroupAnagrams.java
```diff
@@ -0,0 +1,67 @@
...
+	public static void main(String[] args) {
+		String s[] = {"eat", "tea", "tan", "ate", "nat", "bat"};
+		System.out.println("Output:" +groupAnagrams(s));
+	}
+}
+
+
+
+/*--------------------------------------------Using Prime Multiplication--------------------------------------*/
+/*
```
samarthjj, line 39: Please include Time, Space & comments for all the approaches you mention!

src/WordPattern.java
```diff
@@ -0,0 +1,50 @@
+// Time Complexity : O(n): n is number of words or characters in string/pattern
```
samarthjj, line 1: The TC should be O(nk), where k is the length of the longest word in str.
Looking up and adding a string in a hashmap or set is not a constant time operation.

samarthjj:
Good work Akash, please review the inline comments!

samarthjj:
Good work Akash, please review the inline comments!

PR #223, opened 2020-07-14
Completed 3 problems in Hashing1

alekhya-12:
- Please Review
-Thanks.

### DIFF
--- a/WordPattern.java
+++ b/WordPattern.java
@@ -1,0 +1,55 @@
+/**
+Problem: Word Pattern.
+
+Time Complexity: O(n) or O(m), depending on whichever is large; where n is the length of the pattern and m is the length of the string str
+//because initially we check if they're not isomorphic by checking the lengths of both string and pattern
+
+
+
+Space Complexity: O(n), as we use a hashmap, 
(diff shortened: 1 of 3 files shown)


WordPattern.java
```diff
@@ -0,0 +1,55 @@
+/**
+Problem: Word Pattern.
+
+Time Complexity: O(n) or O(m), depending on whichever is large; where n is the length of the pattern and m is the length of the string str
```
samarthjj, line 4: The time complexity should be O(nk), where n is max(len(patern), len(str)), where k is the length of the longest word in str.
Adding and looking up a String in a Hashmap or Hashset is a linear time, not constant, operation

WordPattern.java
```diff
@@ -0,0 +1,55 @@
...
+//because initially we check if they're not isomorphic by checking the lengths of both string and pattern
+
+
+
+Space Complexity: O(n), as we use a hashmap,
```
samarthjj, line 9: The SC should be O(1), since the number of unique chars in pattern cannot exceed 26

samarthjj:
Good job Alekhya, Please review the inline comments!

samarthjj:
Good job Alekhya, Please review the inline comments!

PR #234, opened 2020-07-16
completed all.

satishsolanki1990:


2_isomorphic.py
```diff
@@ -0,0 +1,37 @@
+# time complexity = O(n)
+# space complexity = O(n)
```
samarthjj, line 2: Space is O(1) because the number of unique chars in s or t won't exceed 26.

3_pattern.py
```diff
@@ -0,0 +1,37 @@
+# time complexity = O(n)
```
samarthjj, line 1: O(nk), where n is the number of words in the string and k is the average length of the words.

samarthjj:
Satish, Please review the inline comments. Please include your approach using comments!
Use the template from sample.java file to include your Time & Space analysis, define your variables when mentioning time & space. ex  O(n), where n is the length of s or t.

samarthjj:
Satish, Please review the inline comments. Please include your approach using comments!
Use the template from sample.java file to include your Time & Space analysis, define your variables when mentioning time & space. ex  O(n), where n is the length of s or t.

PR #238, opened 2020-07-19
completed, please review

JaneJianLee:


solution2.py
```diff
@@ -0,0 +1,30 @@
...
+        hm = dict()
+        hm1 = dict()
+        if len(s) != len(t):
+            return False
+
+        for i in range(len(s)):
+            if s[i] in hm:
+                if hm[s[i]] != t[i]:
+                    return False
+            else:
```
samarthjj, line 21: There is no need to make 2 passes over s and t separately, All the operations can be done in a single pass.

solution3.py
```diff
@@ -0,0 +1,30 @@
+#https://leetcode.com/problems/word-pattern/
+
+# // Time Complexity : o(n)
```
samarthjj, line 3: O(n*k), looking up and inserting strings in a hashmap is not a constant operation.

samarthjj:
Good work Jane, Please review the inline comments!
Also please add your 3-4 bullet points of approach for every question!

samarthjj:
Good work Jane, Please review the inline comments!
Also please add your 3-4 bullet points of approach for every question!

PR #245, opened 2020-08-24
Hashing-1 completed Please review

dcarao:


Himanshu-Agarwal93:
Good job. Normally we do not use containsValue() because it is not hashed. Using containsValue() is equivalent to linear searching which defeats our purpose of hashing.

PR #251, opened 2020-08-24
Done Hashing1 - Please review

ramya-sruthi-d:


Himanshu-Agarwal93:
All the problems are from leetcode. Problem 2 is incorrect. Rest problems are correct.

PR #252, opened 2020-08-24
Completed

manicha321:
Please read my comment on Isomorphic problem if you guys test on leetcode. Thanks!


khushal1198:
In the isomorphic string problem you will have to use double mapping to get the correct output. Please correct the solution and upload it again.

PR #272, opened 2020-08-25
Completed Hashing 1 - Please review

mettury:


Anagrams.java
```diff
@@ -0,0 +1,39 @@
+// Time Complexity :O(n)
```
samarthjj, line 1: O(nk), where n is the number of strings in the input and k is the max length of any of the strings.

Isomorphic.java
```diff
@@ -0,0 +1,48 @@
+// Time Complexity :O(n)
+// Space Complexity :O(n)
```
samarthjj, line 2: O(1), Only 26 distinct chars in the hash-map in the worst case.

WordPattern.java
```diff
@@ -0,0 +1,33 @@
+// Time Complexity :O(n)
+// Space Complexity :O(n)
```
samarthjj, line 2: Time & Space both O(nk), where n is the number of strings in the input and k is the max length of any of the strings.
Adding a string to a hashmap does not take constant time. 
The strArr is of size O(nk), the length of the strings have to be taken into account as well.

samarthjj:
Good job Rohit, Please review the inline comments.
Please add a brief description of your approach in 3-4 points!

samarthjj:
Good job Rohit, Please review the inline comments.
Please add a brief description of your approach in 3-4 points!

PR #277, opened 2020-08-25
Completed Hashing-1. Please Review

owaiskazi19:


Isomorphic-String.py
```diff
@@ -0,0 +1,35 @@
...
+            return False
+        
+        for i in range(0, len(s)):
+            ss = s[i]
+            tt = t[i]
+
+            ss = 'first'+ss
+            tt = 'second'+tt
+
+            if ss not in res:
```
Himanshu-Agarwal93, line 26: Looks like you made a mistake in copying the solution. Initialization of `res` is missing.

Himanshu-Agarwal93 (commented):
Good work!

Himanshu-Agarwal93:
Good work!

Isomorphic-String.py
owaiskazi19: Left res here but added it on LC

PR #279, opened 2020-08-25
Completed Hashing 1---Please Review

harinimadhavaram:


GroupAnagrams.java
```diff
@@ -0,0 +1,53 @@
+// Time Complexity : O(n * l) where l is the length of the string
```
samarthjj, line 1: please define n, please define L more specifically. L here should be the maximum length of a string in the input.

Isomorphic.java
```diff
@@ -0,0 +1,47 @@
+// Time Complexity : O(N)
+// Space Complexity :O(M) where M is the length of the string
```
samarthjj, line 2: Space, O(1). Only 26 distinct chars in the hashmap in the worst case.

WordPatter.java
```diff
@@ -0,0 +1,57 @@
+// Time Complexity : O(N)
+// Space Complexity :O(M) where M is the length of the string
```
samarthjj, line 2: Time and space here both should be O(nk), where n is the (number of the words in str)/(length of pattern) and k is the average length of all the words in str.
Adding a string to hashmap does not take constant time. The hashmaps have to take into account the length of the words as well.

samarthjj:
Good job Harini, Please review the inline comments!

samarthjj:
Good job Harini, Please review the inline comments!

PR #284, opened 2020-08-26
Completed Hashing-1 Problems - Please review

sahanaramesh09:


Problem1.java
```diff
@@ -0,0 +1,38 @@
+
+//Problem: leetcode 49. Group Anagrams: Given an array of strings, group anagrams together.
+//Problem Link: https://leetcode.com/problems/group-anagrams/
+// Time Complexity : O(n)
```
samarthjj, line 4: O(n* k log(k)), Sorted every string!

Problem2.java
```diff
@@ -0,0 +1,56 @@
+//Problem: leetcode 205. Isomorphic Strings. 
+//Problem Link: https://leetcode.com/problems/isomorphic-strings/
+// Time Complexity : O(n)
+// Space Complexity :O(2*l)
```
samarthjj, line 4: Space here is O(1), only 26 distinct chars in any hashmap in the worst case.

Problem3.java
```diff
@@ -0,0 +1,48 @@
+//Problem: leetcode 290. word Pattern. 
+//Problem Link: https://leetcode.com/problems/word-pattern/
+// Time Complexity : O(n)
+// Space Complexity :O(n*n)
```
samarthjj, line 4: Time & Space here both are O(nk), where n is the (number of words is str)/(length of pattern) and k is the average length of the strings is str.
Adding a string to a hashmap does not tale constant time. The hashmaps have to take into account the length of the strings.

samarthjj:
Good job Sahana, Please review the inline comments!

samarthjj:
Good job Sahana, Please review the inline comments!

PR #287, opened 2020-08-26
Completed Day5 Problems : Please review (especially time and space complexity)

ShaivyT:


Problem3.java
```diff
@@ -0,0 +1,33 @@
+// Time Complexity : O(n)
+// Space Complexity : O(n)
```
samarthjj, line 2: Both time and space here would be O(nk), where n is the (number of words in str)/(length of pattern), k is the average length of the words in str.
Adding a string to a hashmap does not take constant time. The hashmaps have to take into account the length of the strings as well.

samarthjj:
Good job Shaivy, Please review the inline comments!
Please add the Time & Space complexity and comments to describe your approach as well!

samarthjj:
Good job Shaivy, Please review the inline comments!
Please add the Time & Space complexity and comments to describe your approach as well!

PR #290, opened 2020-08-27
Please review the submission

sidmal11:


Anagram.java
```diff
@@ -0,0 +1,66 @@
+// Time Complexity : O(n)+O(n) since we have to traverse all the strings in the list and then return the values of all the strings by traversing hashmap.values
+// Space Complexity : since we are using a hashmap therefore the worst possible case is that it has to create a hashmap of all the strings in the list anda integer array of 26 so O(n+n)
```
samarthjj, line 2: Time & Space both O(n*k), where is the number of words and k is the average length of the words.

Isomorphic.java
```diff
@@ -0,0 +1,52 @@
+// Time Complexity : O(n) + O(n), where n is the length of the string. Beacuse of the for loop and contain values
+// Space Complexity : The keys and values are character so O(n) + O(n) where n is the length of character.
```
samarthjj, line 2: Time: O(n) only because in this case contains checks for a **character** in a hash-map which is a constant time operation.
Space: O(1), only 26 distinct chars in hash-map in the worst case.

Pattern.java
```diff
@@ -0,0 +1,52 @@
+// Time Complexity : O(n) + O(n), where n is the length of the string. Beacuse of the for loop and contain values
+// Space Complexity : The key is character and value is string. it will be O(n) + O(n*l) where n is the number of characters and l is the length of string
```
samarthjj, line 2: Time & Space both O(n*k), n is the (number of words)/(length of pattern) and k here is the average length of the words.
Adding strings to a hashmap is not constant time. The hashmaps should account for the length of the words as well.

samarthjj:
Good work Siddharth, Please review the inline comments. 
Please add a brief description of your approach for all the problems as well.

samarthjj:
Good work Siddharth, Please review the inline comments. 
Please add a brief description of your approach for all the problems as well.

Pattern.java
sidmal11: Okay I understood. So if i have to add a list of strings in the value of my hashmap. Will Time and space complexity be the length of all the strings in the list?

PR #291, opened 2020-08-27
completed hashing-1

varsha-varadarajan:


IsomorphicStrings.java
```diff
@@ -0,0 +1,33 @@
+/*
+LC 205: Medium: https://leetcode.com/problems/isomorphic-strings/submissions/
+Time Complexity : O(n)
+Space Complexity : O(2n)
```
samarthjj, line 4: O(1), only 26 distinct chars in the hashmap in the worst case.

WordPattern.java
```diff
@@ -0,0 +1,40 @@
+/*
+LC 205: Medium: https://leetcode.com/problems/isomorphic-strings/submissions/
+Time Complexity : O(n)
+Space Complexity : O(n*l) + O(number_of_unique_words); n - no of characters in the pattern, l - average length
```
samarthjj, line 4: Time here will be O(n*l) as well. Adding strings to a hashmap does not take constant time.

samarthjj:
Good job Varsha, Please review the inline comments!

samarthjj:
Good job Varsha, Please review the inline comments!

PR #307, opened 2020-09-19
Completed, please review

siddhesh25:


Problem1.py
```diff
@@ -0,0 +1,19 @@
+# Time Complexity: O(N)
```
Lohith-Aswathappa, line 1: check time complexity
sorting has to be considered

Lohith-Aswathappa (commented):
Good work, Siddhesh!
Check the inline comment

Lohith-Aswathappa:
Good work, Siddhesh!
Check the inline comment

PR #314, opened 2020-09-19
Completed Problem 1,2,3

AkashBudholia:


Problem1.java
```diff
@@ -0,0 +1,59 @@
+// Time Complexity : O(n)
```
Lohith-Aswathappa, line 1: consider sorting into time complexity

Lohith-Aswathappa (commented):
Good work, Akash!
Check inline comment

Lohith-Aswathappa:
Good work, Akash!
Check inline comment

Problem1.java
AkashBudholia: Yes, my bad, it would be nlogn

PR #331, opened 2020-09-23
hashing1

Lakshya0744:


anotate_group.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+dict={}
+
+# print(dict)
+
+
+for i in input:
+    store = (''.join(sorted(i)))
+    if dict[store]:
```
kannan-mrugapphan, line 13: Please try another approach without sorting

kannan-mrugapphan (commented):
Good Job

kannan-mrugapphan:
Good Job

PR #364, opened 2020-10-21
Solution to Hashing-1

sumitjanawlekar:
I have added some doubts, in the "did you face any problems while solving this question" section. Please address them.

### DIFF
--- a/group_anagrams.java
+++ b/group_anagrams.java
@@ -1,0 +1,48 @@
+// Time Complexity : O(nl), where "n" is the length of the input array and "l" is the length of each string
+// Space Complexity : O(n), where n is the length of input array
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : Just wanted to know, if I consider brute-force solution as I will compare each string
+                                             //with all other strings in the array (O(n^2)), AND at every comparision I will first
+                                             // sort both the strings (O(l logl) where l is the length of each string), and compare
+                                             //if both the string are same (O(l) --> string comparison), save that string against the
+                                             //sorted string's (will be hashmap key) value
+                                             // In this case, does the Time Complexity become O(n^2 (l logl + l)) = O(n^2*l*logl) ???
+
+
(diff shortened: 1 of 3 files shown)


isomorphic.java
```diff
@@ -0,0 +1,49 @@
+// Time Complexity : O(n), where n is the length of the string s and t
+// Space Complexity :O(1), though we are using hashMap, but at max it will store (all uppercase and lowercase), which is constant space
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : Was a bit confused with the space complexity, please correct me if I am wrong
```
kannan-mrugapphan, line 4: Yeah you are right. Its O(1) as in worst case all unicode chars can be stored in the map

word_pattern.java
```diff
@@ -0,0 +1,56 @@
+// Time Complexity :O(n)
+// Space Complexity :O(n) where n is the length of String pattern and String s (because a new array will be created for storing String s as array of Strings)
+// Did this code successfully run on Leetcode :
+// Any problem you faced while coding this : I am confused on the space complexity, because in the isomorphic problem, we said
```
kannan-mrugapphan, line 4: If you observe, the key in the second map is the actual word(string). It depends on the input and so its O(n). The space can't be reduced to constant.

group_anagrams.java
```diff
@@ -0,0 +1,48 @@
+// Time Complexity : O(nl), where "n" is the length of the input array and "l" is the length of each string
+// Space Complexity : O(n), where n is the length of input array
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : Just wanted to know, if I consider brute-force solution as I will compare each string
```
kannan-mrugapphan, line 4: Yeah this analysis is right.

kannan-mrugapphan (commented):
Good Job

kannan-mrugapphan:
Good Job

PR #387, opened 2020-11-27
Hashing-1 done

snk95:


Group anagrams togegther.java
```diff
@@ -0,0 +1,63 @@
...
+            
+            hashmap[product].append(s)
+            
+        
+        return hashmap.values()
+
+*/
+
+// Time Complexity : O(nk) where n is size of array and k is size of largest string
+// Space Complexity : O(1) as alphabets are only 26
```
samarthjj, line 28: Space: O(nk), The Hashmap is storing words as values. 
n: Number of words, k: Average length of words.

samarthjj:
Good work Sarvesh, Please review the inline comments!

samarthjj:
Good work Sarvesh, Please review the inline comments!

Group anagrams togegther.java
snk95: I think it should be O(n) as I am storing product of prime numbers as key and values are words

PR #415, opened 2020-12-29
Completed- hashing-1

shamali21:


ramya-sruthi-d (commented):
Code is fine, the tc for problem 1 is wrong because you are sorting the string. Also, mention both tc and sc for all the problems.

ramya-sruthi-d:
Code is fine, the tc for problem 1 is wrong because you are sorting the string. Also, mention both tc and sc for all the problems.

PR #417, opened 2020-12-29
solving brute force

yahska-ednihs:
Will add comments after class


samarthjj:
Good work Akshay, Please also try all the problems the way they were done in class.

samarthjj:
Good work Akshay, Please also try all the problems the way they were done in class.

PR #422, opened 2020-12-29
Hashing-1 Done, Please check

Sowmya246:


isomorphic.py
```diff
@@ -0,0 +1,23 @@
+Time Complexity: O(NKlogK)
```
Lohith-Aswathappa, line 1: could you check the time complexity again. 
I believe the TC should be O(N) according to the logic

wordPattern.java
```diff
@@ -0,0 +1,37 @@
+Time Complexity: O(NlogN)
```
Lohith-Aswathappa, line 1: check the TC here 
It should be O(N), also if I'm wrong please let me know

Lohith-Aswathappa (commented):
Good work, Sowmya!
Check the inline comments

Lohith-Aswathappa:
Good work, Sowmya!
Check the inline comments

PR #429, opened 2020-12-30
Completed Hashing-1

Stnagam:


Problem1.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.*;
+class Problem1{
+
+    
+    public List<List<String>> groupAnagrams(String[] strs) {
+        List<List<String>> res = new ArrayList<>();
+        HashMap<String, List<String>> map = new HashMap<>();
+        for (String word : strs) {
+            char wordArr[] = word.toCharArray();
+            Arrays.sort(wordArr);
```
Lohith-Aswathappa, line 20: You are sorting every word in this logic, this has to be considered in the time complexity
TC would be n MlogM -> n words and MlogM sorting

Lohith-Aswathappa (commented):
Good work, Sai!
Check the inline comments

Lohith-Aswathappa:
Good work, Sai!
Check the inline comments

PR #432, opened 2020-12-30
Please Review: Isomorphic String

Manasi2020:
This problem can also be solved using Char array and their ascii values which is constant space.
In this problem we cannot use one HashMap as it won't verify If the strings are isomorphic other way around


ramya-sruthi-d (commented):
Good work, mention the complexities of second problem as well.

ramya-sruthi-d:
Good work, mention the complexities of second problem as well.

PR #459, opened 2021-01-25
Assignment Hashing1 Please Review!

srivastavavertika201:


2 strings are Isomorphic
```diff
@@ -0,0 +1,19 @@
...
+        for(int i = 0;i<s.length();i++){
+            hashSet1.add(s.charAt(i));
+            hashSet2.add(t.charAt(i));
+        }
+        if(hashSet1.size()==hashSet2.size()){
+            return true;
+        }
+        return false;
+        
+    }
```
Lohith-Aswathappa, line 14: Try using a HashMap and map the character that can be replaced to find the isomorphic words

Word Pattern
```diff
@@ -0,0 +1,21 @@
...
+            if(hm.containsKey(pattern.charAt(i))){
+                if(hm.get(pattern.charAt(i)).equals(arr[i])){
+                    flag = true;
+                }
+                else
+                    return false;
+            }
+            hm.put(pattern.charAt(i),arr[i]);
+        } 
+        return flag;
```
Lohith-Aswathappa, line 17: Logic is correct 
Implement the same logic considering the pattern as a main string and mapping it to character

Lohith-Aswathappa (commented):
Nice try, Vertika!
Check the comments and update the code

Lohith-Aswathappa:
Nice try, Vertika!
Check the comments and update the code

PR #494, opened 2021-02-22
Completed Hashing-1

babzi17:


Problem2.java
```diff
@@ -0,0 +1,23 @@
...
+            }
+            
+            mapS.put(s.charAt(i), i);
+            mapT.put(t.charAt(i), i);
+        }
+        return true;
+    }
+}
+//time - O(N)
+//space - O(N)
```
Lohith-Aswathappa, line 23: space would be o(1) since there are only 26 characters

Lohith-Aswathappa (commented):
Good work, Vinay!
Check inline comment

Lohith-Aswathappa:
Good work, Vinay!
Check inline comment

PR #500, opened 2021-02-22
Problem 2

SourabhSanghi:
Isomorphic strings


Problem 2
```diff
@@ -0,0 +1,30 @@
+class Solution {
+    public boolean isIsomorphic(String s, String t) {
```
yahska-ednihs, line 2: Good Job! with this problem. Please add all the information requested with the solution. Also Please add all the other solutions as well.

PR #501, opened 2021-02-22
Problem 3

SourabhSanghi:
Word-Pattern


Problem 3
```diff
@@ -0,0 +1,26 @@
+class Solution {
+    public boolean wordPattern(String pattern, String s) {
+        String[] sArray = s.split(" ");
```
yahska-ednihs, line 3: Good Job! Please add comments explaining your approach and all the information requested for the solution.

PR #502, opened 2021-02-22
Problem 1

SourabhSanghi:
Please review. Not working for 1 test case , as per leet code

### DIFF
--- a/Problem 1
+++ b/Problem 1
@@ -1,0 +1,45 @@
+class Solution {
+    public List<List<String>> groupAnagrams(String[] strs) {
+        
+        HashMap<Long , List<String>> stringMap = new HashMap<>();
+        
+        // loop through each element in input array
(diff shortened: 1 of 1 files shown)


Problem 1
```diff
@@ -0,0 +1,45 @@
...
+            stringMap.get(primeVal).add(s);
+        }
+        return new ArrayList<>(stringMap.values());
+    }
+        
+        // Calculate  numeric value of a string
+        private long calculateNumericValue(String s){
+           
+            int[] primeNumbers = {2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83,                                     89, 97, 101};
+            int ans = 1;
```
sanchitpruthi, line 22: This should be long instead of int -> overflow error

PR #511, opened 2021-02-23
Completed Hashing-1.

Tirthaj:


Problem 1.py
```diff
@@ -0,0 +1,12 @@
...
+# Space Complexity: O(nk)
+from collections import defaultdict
+
+
+class Solution:
+    def groupAnagrams(self, strs: List[str]) -> List[List[str]]:
+        dic = defaultdict(list)
+
+        for i in strs:
+            dic["".join(sorted(i))].append(i)
```
Lohith-Aswathappa, line 11: sorting has to be considered for the time complexity. O(NKlogK) where N is the number of Strings and KlogK for sorting

Problem 3.py
```diff
@@ -0,0 +1,32 @@
+# Time Complexity : O(n)
+# Space Complexity : O(1)
```
Lohith-Aswathappa, line 2: dictionary has to considered for the space. 
Character dict will be constant but the other dictionary varies

Lohith-Aswathappa (commented):
Good work, Tirthaj!
Check inline comments

Lohith-Aswathappa:
Good work, Tirthaj!
Check inline comments

PR #527, opened 2021-03-23
Completed Hashing 1

Nitishmane:
Completed Hashing 1


maithrreye:
Good Job Nitish ! Please add TC, SC and comments for the problem.

maithrreye:
Good Job Nitish ! Please add TC, SC and comments for the problem.

PR #584, opened 2021-05-31
Completed Hashing-1. Please Review.

RadhikaKulk:
Group Anagrams
- TC: O(n * mlogm) where n are total strings in the input array, and m is the size of each string (mlogm becasue I am sorting)
- SC: O(n)

Isomorphic Strings
- TC: O(n) where n is the length of the string
- SC: O(n)

Word Pattern
- TC: O(n) where n is the number of words in s OR n is length of pattern
- SC: O(n)

### DIFF
--- a/GroupAnagrams.java
+++ b/GroupAnagrams.java
@@ -1,0 +1,27 @@
+class Solution {
+    public List<List<String>> groupAnagrams(String[] strs) {
+        HashMap<String, List<String>> map = new HashMap<>();
+        char[] temp;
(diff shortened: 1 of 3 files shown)


Rooppesh:
Good Job! Please add the Time and Space Complexities in the respective files, so that we get a clear picture for each approach.

Rooppesh:
Good Job! Please add the Time and Space Complexities in the respective files, so that we get a clear picture for each approach.

PR #633, opened 2021-06-03
Hashing -1 

hharshitarora:


hharshitarora:
For this my main doubt was regarding the last word pattern problem. Overall I used the same idea as the isometric part, but my main concern was that I made an arraylist to make the list for words and then i do the same problems.

hharshitarora:
Would really appreciate some thought on the above and its correctness.

PR #644, opened 2021-06-06
Hashing-1 completed

kk25081998:


maithrreye:
Good Job ! The TC of Isomorphic string and word pattern is O(1) because you are searching using Key - value in dict ! and the space compelxity is 0(1) as well as the dictionary are all Input and you didnt have auxillary space.

maithrreye:
Good Job ! The TC of Isomorphic string and word pattern is O(1) because you are searching using Key - value in dict ! and the space compelxity is 0(1) as well as the dictionary are all Input and you didnt have auxillary space.

PR #668, opened 2021-06-28
Completed Hashing - 1: Please Review

ankitachikodi:


GroupAnagrams.java
```diff
@@ -0,0 +1,26 @@
+// Time Complexity: O(N)
```
super30admin, line 1: We are sorting the strings also. So klogk term will be extra where k is average length of string

IsomorphicStrings.java
```diff
@@ -0,0 +1,31 @@
...
+        for(int i=0; i<s.length(); i++) {
+            Character st = s.charAt(i);
+            Character ts = t.charAt(i);
+            
+            if(map.containsKey(st)) {
+                 if(ts != map.get(st)) {
+                     return false;
+                 } 
+            } else {
+                if(!map.containsValue(ts))
```
super30admin, line 21: This will turn the solution in O(n2). Look up on values in map is O(n). This is not a good solution.

WordPattern.java
```diff
@@ -0,0 +1,32 @@
...
+        HashMap<Character, String> map = new HashMap<Character, String>();
+
+        for(int i=0; i<arr.length; i++){
+            char c = pattern.charAt(i);
+
+            if(map.containsKey(c)){
+                if(!map.get(c).equals(arr[i]))
+                    return false;
+            }else{
+                if(map.containsValue(arr[i]))
```
super30admin, line 23: Again Look up on values should not be done. Make a separate map

super30admin (commented):
Read inline feedback

PR #806, opened 2021-10-19
Completed Anagram problems

Jeevitha-web:


brunda4:
Please provide the correct SC, for the problem.

PR #832, opened 2021-12-01
solved hashing 1. please review

akulprabhudesai:


brunda4 (commented):
Isomorphic strings has a linear SC due to use of the Hashmap.

brunda4:
Isomorphic strings has a linear SC due to use of the Hashmap.

akulprabhudesai:
Hello,
I think HashMap will have constant space complexity for following reasons.
Please correct me if I am wrong:

1. In HashMap only fixed no of entries will go which is equal to total no
of ascii characters and hence how much data HashMap has or will be having
is not dependant on input.
2. So irrespective of how much chars input strings S and T has, I know how
many entries HashMap will have even in worst case which is equal to total
no of valid ascoi chars.
3. Hence HashMap will not contribute towards SC in this peoblem.


Please let me know if there are any corrections to be made.


Waiting for your reply

Thank you

PR #867, opened 2021-12-28
submit hashing1

hitesh-balapanuru:


manojvarmarudraraju (commented):
Good job, In isomorphics problem key in char_dict.keys() is O(1) but char_dict.values() will take O(N) time. So it is better to use two dictionaries!!

manojvarmarudraraju:
Good job, In isomorphics problem key in char_dict.keys() is O(1) but char_dict.values() will take O(N) time. So it is better to use two dictionaries!!

PR #925, opened 2022-02-08
Completed Hashing-1 - Please Review

suhaas-livcd:
- Used same approach for Isomorphic and WordPattern, need to optimise logic using prime nos or ascii values


shrivatsav:
Good Work!

PR #974, opened 2022-03-12
Please check Hashing-1

shrutikamokashi:
For some reason the code for Problem 3 fails, it isn't working and returning false in "if (pMap.get(pChar) != sStr)" condition.
Please check it and help me.

### DIFF
--- a/Problem1.java
+++ b/Problem1.java
@@ -1,0 +1,63 @@
+import java.util.*;
+//import java.util.HashMap;
+
+// Time Complexity
+// With bruteforce solution: O(nk log k)
+// With hashing: O(n)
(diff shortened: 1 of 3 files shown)


keshav-Gupta-IS (commented):
Problem 3 that error can be resolved using .equals method but your logic will also fail for cases like s: "aaa"  p: "aa aa aa aa"!! So I would suggest go through the problem again and rework the logic as well.

keshav-Gupta-IS:
Problem 3 that error can be resolved using .equals method but your logic will also fail for cases like s: "aaa"  p: "aa aa aa aa"!! So I would suggest go through the problem again and rework the logic as well.

PR #1028, opened 2022-05-04
Completed Hashing-1 problems. Please Review

SaurabhGujare:
Problem 1:
Given an array of strings, group anagrams together.

Problem 2:
Given two strings s and t, determine if they are isomorphic. 

Problem 3:
Given a pattern and a string str, find if str follows the same pattern. Here follow means a full match, such that there is a bijection between a letter in pattern and a non-empty word in str.

### DIFF
--- a/GroupAnagrams.java
+++ b/GroupAnagrams.java
@@ -1,0 +1,24 @@
+// Time Complexity : O(N.NLogN), O(N) for outer loop O(NlogN) for sorting
+// Space Complexity :O(NK),N=arr.length, K=max(strs.length)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No
+import java.util.*;
+
+public class GroupAnagrams {
+    public List<List<String>> groupAnagrams(String[] strs) {
+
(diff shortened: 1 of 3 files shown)


GroupAnagrams.java
```diff
@@ -0,0 +1,24 @@
+// Time Complexity : O(N.NLogN), O(N) for outer loop O(NlogN) for sorting
```
shrivatsav, line 1: Please check the time complexity.

IsIsomorphic.java
```diff
@@ -0,0 +1,30 @@
+// Time Complexity : O(N)
+// Space Complexity : O(1), size is  fixed, do not change with input
```
shrivatsav, line 2: Please specify what `N` indicates.

shrivatsav (commented):
Good Work!

shrivatsav:
Good Work!

IsIsomorphic.java
SaurabhGujare: N is the length of each string. Strings of equal lengths

PR #1058, opened 2022-05-10
Hashing 1

arshmohd:
1. Group Anagrams implementation using HashMap.
2. Word pattern which is kind of simialr to Anagrams concept its a word implementation using HashMap.
3. Isomorphic String check implementation using HashMap.

### DIFF
--- a/GroupAnagrams.java
+++ b/GroupAnagrams.java
@@ -1,0 +1,45 @@
+
+//Time Complexity:  O(n\times klogk)
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


GroupAnagrams.java
```diff
@@ -0,0 +1,45 @@
+
+//Time Complexity:  O(n\times klogk)
```
shrivatsav, line 2: Please specify what `n` and `k` indicates. Please include the space complexity.

shrivatsav (commented):
Good Work!

shrivatsav:
Good Work!

PR #1107, opened 2022-06-03

akshaykmus:


Isomorphic.java
```diff
@@ -0,0 +1,33 @@
+
+// Time Complexity : O(1)
+// Space Complexity : O(1)
```
Sumesh-H, line 3: How?

Sumesh-H:
Good Job! Reconsider the space complexity for the Isomorphic problem. Also, try the other problem as well.

Sumesh-H:
Good Job! Reconsider the space complexity for the Isomorphic problem. Also, try the other problem as well.

Isomorphic.java
akshaykmus: At the most we store O(52) which is equal to O(1) right?

Sumesh-H: You have used an auxiliary data structure i.e. Hashmap in your approach. It takes O(n) space complexity.

PR #1120, opened 2022-06-08
All the problems are solved please review

Dhairya1999:


Problem2_isomorphic.py
```diff
@@ -0,0 +1,23 @@
+'''
+Time Complexity : O(1)
```
jeelpatel18, line 2: O(N) is the TC

Problem3_pattern.py
```diff
@@ -0,0 +1,26 @@
+'''
+Time Complexity : O(1)
```
jeelpatel18, line 2: O(N) is the TC

Problem1_anagram.py
```diff
@@ -0,0 +1,19 @@
+'''
+Time Complexity : O(nlogn)
```
jeelpatel18, line 2: Nk * log k will be TC, where k = avg. length of word

jeelpatel18: You sort the each chars of string/word here

jeelpatel18 (commented):
Good Job!!!

jeelpatel18:
Good Job!!!

PR #1132, opened 2022-06-27
Completed Sprint Hashing-Please Review

kunalwvm:


niharikagupta2021 (commented):
Good Job! Please add sc and tc as well

PR #1134, opened 2022-06-28
word-pattern.py

vish1525:


word-pattern.py
```diff
@@ -0,0 +1,30 @@
+"""
+Runtime Complexity:
+O(m*k) - where 'm' is the number of character which maps to 'k' characters in 's' string.
```
ishika2741, line 3: Kindly check the complexity.

ishika2741 (commented):
Good job!

PR #1135, opened 2022-06-28
isomorphic-string.py

vish1525:


isomorphic-string.py
```diff
@@ -0,0 +1,26 @@
+"""
+Runtime Complexity:
+O(n) - because we run a loop to add all the characters in s and t string to newly created hashmaps s_map and t_map.
+O(n) - Space complexity because we create a new hashmap data structure and add 'n' elements to it.
```
ishika2741, line 4: At max , we will be having 26 lower case characters in hashmap . So space complexity will be constant.

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1157, opened 2022-06-28
Hashing One 

SreekarManchukonda:


niharikagupta2021:
Good Job! Please add sc and tc as well

PR #1258, opened 2022-09-01
Completed - Please Review

sid-pandya:


Word Pattern.py
```diff
@@ -0,0 +1,27 @@
+# Time Complexity : O(n + m)
```
daredevilteja, line 1: what is n and m

daredevilteja:
Good Work! Keep it up

PR #1272, opened 2022-09-14
Completed Sprint Hashing-1 - Please Review

jaszaveri43:


Question-2.py
```diff
@@ -0,0 +1,18 @@
+#Time-Complexity: O(n*n)
```
jeelpatel18, line 1: It is not n*n, it will be O(N) only

Question-3.py
```diff
@@ -0,0 +1,22 @@
+#Time-Complexity: O(n*n)
```
jeelpatel18, line 1: Here Also it is not n^2

jeelpatel18 (commented):
Well done

jeelpatel18:
Well done

PR #1287, opened 2022-09-15
Completed Hashing 1 - Pls Review

lakshaynanda:


wordPattern.java
```diff
@@ -0,0 +1,30 @@
+// TC: O(n)
+// SC: O(1)
```
daredevilteja, line 2: SC is wrong

daredevilteja (changes requested):
Please mention what is n,k

daredevilteja:
Please mention if it ran successfully on leetcode or not, for every problem.

wordPattern.java
lakshaynanda: By mistake I wrote O(1). I'll update this.

PR #1334, opened 2022-10-18
hashing1 done

tanaytadas16:


wordPattern.java
```diff
@@ -0,0 +1,29 @@
+public class wordPattern {
```
daredevilteja, line 1: Please mention TC and SC

daredevilteja (changes requested):

daredevilteja:
Good Work! Keep it up

PR #1523, opened 2023-07-05

Hemanth-Reddy98:


Problem1.py
```diff
@@ -0,0 +1,12 @@
+class Solution:
+    # Time Complexity: O(NM log M)
```
ishika2741, line 2: Kindly mention what m and n is

Problem2.py
```diff
@@ -0,0 +1,14 @@
+class Solution:
+    # Time Complexity: O(N)
+    # Space Complexity: O(N)
```
ishika2741, line 3: Space will be constant. O(1) because in worst case max 256 valid ascii characters can be there.

ishika2741 (commented):
Well done!

ishika2741:
Well done!

PR #1531, opened 2023-07-09
Completed Hashing-1

PurnaSahithiAdduri:
Completed Hashing-1

### DIFF
--- a/Problem2.java
+++ b/Problem2.java
@@ -1,0 +1,30 @@
+// Time Complexity : O(n) 
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : no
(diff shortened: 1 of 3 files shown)


Problem2.java
```diff
@@ -0,0 +1,30 @@
+// Time Complexity : O(n) 
+// Space Complexity : O(n)
```
ishika2741, line 2: Check space complexity. Since s and t can have all the valid ascii characters, i.e. 256 so it will be constant

ishika2741 (commented):
Good work! Kindly submit prime product solution as well for the anagrams question .

ishika2741:
Good work! Kindly submit prime product solution as well for the anagrams question .

PR #1570, opened 2023-09-04
Hashing 1 complete

GunjanHardasmalani:


Nandha2Tech (commented):
Great Work!
In group anagrams, Since Arrays.sort is used, TC of k log k would also be there. So TC would be O(n klog k).
In IsIsomorphic problem, at the max only 26 characters will be stored in hashmap. So SC would be O(1).

Nandha2Tech:
Great Work!
In group anagrams, Since Arrays.sort is used, TC of k log k would also be there. So TC would be O(n klog k).
In IsIsomorphic problem, at the max only 26 characters will be stored in hashmap. So SC would be O(1).

PR #1600, opened 2023-10-03
Hashing-1 completed and ready for review

smithaholla:


Problem3.cs
```diff
@@ -0,0 +1,48 @@
+public class WordPattern
+    {
+        // Time Complexity : O(1) -- i think its constant time as the length of the entire string s is 3000 and t is 300
```
ishika2741, line 3: Its not constant. 3000 is not a small no.

ishika2741 (commented):
Good work!

ishika2741:
Good work!

Problem3.cs
smithaholla: Yes thank you.  I realized it after..

PR #1636, opened 2023-11-14
 Solved Hashing-1

yashfirke:


Nandha2Tech (commented):
Great Work! In problem3 try to use Integer datatype to store the index in HashMap. It will pass all test cases.

Nandha2Tech:
Great Work! In problem3 try to use Integer datatype to store the index in HashMap. It will pass all test cases.

PR #1649, opened 2023-12-25
Completed Hashing 1

goodguygoddy:


Problem_1.cpp
```diff
@@ -0,0 +1,38 @@
+/*
+ * Time Complexity : O(n)
```
ishika2741: Check time and space complexity please. Since you are sorting strings here , it would take more time. Also go through the class videos and try to submit optimized solution for this problem.

ishika2741 (commented):
Good work!

ishika2741:
Good work!

goodguygoddy:
Yes, correct the Time Complexity of this approach in Problem 1 would be O(n*k*log(k)) where the n is the length of the list of the words and the k is the average string length of each word.

God7wyn:
Yes, correct the Time Complexity of this approach in Problem 1 would be O(n*k*log(k)) where the n is the length of the list of the words and the  k is the average string length of each word.

PR #1784, opened 2024-06-24
completed hashing-1

dhikki-study:
#############49. Group Anagrams####################
// Time Complexity : N
// Space Complexity : N
// Did this code successfully run on Leetcode : YES
// Any problem you faced while coding this : NA

// Your code here along with comments explaining your approach in three sentences only

// for each string I will sort it alphabetically and check if it present in dict or not if it is not present I will use sorted string as key and populate its position  in strs as value. So now all the string which has same character will be under 1 key along with positions which I will read and put it in 1 list

#############205. Isomorphic Strings####################
// Time Complexity : N
// Space Complexity : N
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : NA

// Your code here along with comments explaining your approach in three sentences only

// in this code also I will use dictionary, I will take 1st element from 2 string and map them in dictionary, so when 1st element is fetched form both string considering s as master string I will check if 1st element from S is present in dict if it is present I will check whether the value corresponding to it same as 1st letter from 't'. If it is not matching return FALSE or else move to next string. In case any key is not present it will be added in dict

##############290. Word Pattern########################
// Time Complexity : N 
// Space Complexity : N
// Did this code successfully run on Leetcode : YES
// Any problem you faced while coding this : no

// Your code here along with comments explaining your approach in three sentences only

// this code is also same approach as "Isomorphic Strings", here instead of second string there are words which I have moved to list

### DIFF
--- a/Sample
+++ b/Sample
@@ -1,0 +1,104 @@
+#############49. Group Anagrams####################
+
+// Time Complexity : N
+// Space Complexity : N
+// Did this code successfully run on Leetcode : YES
+// Any problem you faced while coding this : NA
+
+
+// Your code here along with comments explaining your approach in three sentences only
+
+// for each string I will sort it alphabetically and check if it present in dict or not if it is not present I will use sorted string as key and populate its position  in strs as value. So now all the string which has same character will be under 1 key along with positions which I will read and put it in 1 list
+
+class Solution:
+    def groupAnagrams(self, strs: List[str]) -> List[List[str]]:
+        dict1={}    #dict initialised
+        list2=[]
+        for c,i in enumerate(strs):	# will give position and word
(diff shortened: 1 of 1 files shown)


Nandha2Tech (commented):
Great Work!
TC for group anagrams is since sorting of the words is done if k is the length of the word TC would be (N * k log k). 
For Isomorphic Strings SC would be O(1) since max size of the hashmap would be 26 characters only.

Nandha2Tech:
Great Work!
TC for group anagrams is since sorting of the words is done if k is the length of the word TC would be (N * k log k). 
For Isomorphic Strings SC would be O(1) since max size of the hashmap would be 26 characters only.

PR #1815, opened 2024-06-30
Completed Hashing 1

pradnyabhukan:


Nandha2Tech (commented):
Great Work! SC for isomorphic strings problem would be O(1) since hash map has at most 26 characters.

Nandha2Tech:
Great Work! SC for isomorphic strings problem would be O(1) since hash map has at most 26 characters.

pradnyabhukan:
> Great Work! SC for isomorphic strings problem would be O(1) since hash map has at most 26 characters.

Oh okay. Thanks for clarifying !

PR #1844, opened 2024-08-04
Hashing 1 Changes

botlamanoj29:


Nandha2Tech (commented):
Great Work! SC IsomorphicStrings would be O(1) as there will not be more than 26 characters. TC for anagrams problem would be O(n * k log K) since sorting of each word is happening in your logic.

Nandha2Tech:
Great Work! SC IsomorphicStrings would be O(1) as there will not be more than 26 characters. TC for anagrams problem would be O(n * k log K) since sorting of each word is happening in your logic.

PR #1871, opened 2024-10-13
Completed Hashing1

ravikanthkarra:


Nandha2Tech (commented):
Great Work! Space complexity for IsIsomorphic would be O(1) since map will have atmost 26 characters only.

Nandha2Tech:
Great Work! Space complexity for IsIsomorphic would be O(1) since map will have atmost 26 characters only.

PR #1875, opened 2024-10-28
done hashing 1

komalnagpal:


Nandha2Tech (commented):
Great Work! Space complexity for IsIsomorphic strings would be O(1) since hashmap will not have more than 26 characters.

Nandha2Tech:
Great Work! Space complexity for IsIsomorphic strings would be O(1) since hashmap will not have more than 26 characters.

PR #1889, opened 2024-11-03
Hashing-1 Completed

pankhuri0209:


Nandha2Tech (commented):
Great Work! Space complexity for isIsomorphic would be O(1) since the size of the map will not exceed after 26 characters.

Nandha2Tech:
Great Work! Space complexity for isIsomorphic would be O(1) since the size of the map will not exceed after 26 characters.

PR #1997, opened 2025-02-01
Done Hashing-1

RajatOnkar:


Nandha2Tech (commented):
Great Work! In problem2, when you convert hashmap to char strings, there are some issues in the way how you handle the list based on index. Please refer the below code to correct the issues
class Solution(object):
    def isIsomorphic(self, s, t):
        """
        :type s: str
        :type t: str
        :rtype: bool
        """
        s_len = len(s)
        t_len = len(t)
        if s_len != t_len: return False
        sChar = [None] * 100
        tChar = [None] * 100
        for i in range(s_len):
            s_character = s[i]
            t_character = t[i]

            if sChar[ord(s_character) - ord(" ")] is None:
                sChar[ord(s_character) - ord(" ")] = t_character
            else:
                if sChar[ord(s_character) - ord(" ")] != t_character:
                    return False

            if tChar[ord(t_character) - ord(" ")] is None:
                tChar[ord(t_character) - ord(" ")] = s_character
            else:
                if tChar[ord(t_character) - ord(" ")] != s_character:
                    return False
        return True

Nandha2Tech:
Great Work! In problem2, when you convert hashmap to char strings, there are some issues in the way how you handle the list based on index. Please refer the below code to correct the issues
class Solution(object):
    def isIsomorphic(self, s, t):
        """
        :type s: str
        :type t: str
        :rtype: bool
        """
        s_len = len(s)
        t_len = len(t)
        if s_len != t_len: return False
        sChar = [None] * 100
        tChar = [None] * 100
        for i in range(s_len):
            s_character = s[i]
            t_character = t[i]

            if sChar[ord(s_character) - ord(" ")] is None:
                sChar[ord(s_character) - ord(" ")] = t_character
            else:
                if sChar[ord(s_character) - ord(" ")] != t_character:
                    return False

            if tChar[ord(t_character) - ord(" ")] is None:
                tChar[ord(t_character) - ord(" ")] = s_character
            else:
                if tChar[ord(t_character) - ord(" ")] != s_character:
                    return False
        return True

PR #2115, opened 2025-06-24
Hashing-1 problem

mdashora:


super30admin:
Strengths:
- The solution correctly implements the logic to check for isomorphic strings using two hash maps, ensuring both forward and backward mappings are consistent.
- The time and space complexity are optimal for the problem.

Areas for Improvement:
- The problem title and description in the student's solution do not match the reference solution's problem (Hashing-1 vs. Isomorphic Strings). Ensure the problem context is clear when submitting solutions.
- While the space complexity is O(1), it might be helpful to explicitly mention that the hash maps are bounded by the size of the character set (26 for English letters) to clarify why it's considered O(1).

PR #2119, opened 2025-06-25
Completed Hashing 1

sethu95:


super30admin:
Strengths:
- The code is clean, modular, and well-commented, making it easy to understand.
- The student correctly analyzes and states the time and space complexity for each solution.

Areas for Improvement:
- In Problem1, the use of a char array initialized to '0' and then incremented is a bit unconventional. It would be more straightforward to use an int array initialized to 0, as the frequency counts are naturally integers.
- For Problem2 and Problem3, while the solutions are correct, it might be worth mentioning that the space complexity is technically O(1) for Problem2 because the number of unique characters is fixed (assuming ASCII or Unicode), but this is a minor point.

PR #2127, opened 2025-06-26
Completed Hashing-1

joelkny97:


super30admin:
- The student has done an excellent job in implementing the solutions correctly and efficiently. The code is well-commented and easy to understand.
- For Exercise_1.py, the student could consider removing the commented-out code for the first two methods to keep the final submission clean, as they are not used in the final solution.
- The use of lambda in defaultdict for Exercise_2.py is a good practice, but it could be simplified using a regular dictionary with a default value of None.
- The solution for Exercise_3.py is concise and clever, but it might benefit from a brief explanation of why the condition works (i.e., the bijection between pattern and words).

PR #2137, opened 2025-06-29
Completed Hashing-

devashri-m:


super30admin:
- Problem1.py: Your solution is correct and follows a common approach for grouping anagrams. However, the time complexity could be improved by using a hashing method like the reference solution, which avoids sorting each string. Consider exploring alternative hashing techniques for better performance.
- Problem2.py and Problem3.py: Both solutions are well-implemented and efficient. The use of two dictionaries for bidirectional mapping is a good practice. The code is clean and easy to understand.
- General: Your code is well-structured and follows best practices. Keep up the good work! For future problems, consider exploring more efficient algorithms when possible, especially for problems like grouping anagrams where multiple approaches exist.

PR #2149, opened 2025-07-06
Done Hashing-1

SimranAtre:


super30admin:
- Problem1.py: Fix the syntax error in `sortedstr= '',join(sorted(i))` by removing the extra comma. Consider using a prime number hash for better efficiency.
- Problem2.py: Well done! The solution is correct and efficient.
- Problem3.py: Move the length check `if len(s) != len(pattern):` outside the loop to avoid unnecessary checks. The rest of the solution is correct and efficient.

Overall, your solutions are mostly correct and efficient, but there are minor issues to address. Pay attention to syntax and logical flow to avoid errors.

PR #2173, opened 2025-08-11
Completed

sowbaranika1302:


super30admin:
- Strengths:
  - The use of two hashmaps in prob2.py and prob3.py to ensure bijective mapping is a good approach.
- Areas for Improvement:
  - In prob1.py, while the solution is correct, the reference solution uses a prime number hashing technique which might be more efficient in some cases. The student's approach is still valid and efficient.
  - The comments in prob1.py are slightly cut off and could be more concise.

PR #2178, opened 2025-08-25
Update README.md

AsiaOsman06:


super30admin:
- Strengths: The solution is correct and well-implemented. The code is clean, readable, and follows good practices. The explanation at the top is concise and helpful.
- Areas for Improvement: The time complexity could be improved by using a more efficient hashing method, such as the prime number multiplication approach shown in the reference solution. This would avoid the O(K log K) sorting step for each string. Additionally, the student should ensure that the time complexity notation is accurate (e.g., "O(n.m)" should be clarified to "O(N * K log K)").

PR #2196, opened 2025-08-31
Create isomorphic_strings

Harsha-Ramesh-Naik:


super30admin:
Your solution for checking isomorphic strings is well-written and efficient. You've correctly implemented the one-to-one character mapping check using a dictionary and a set to track used characters. The code is clean and easy to understand. 

However, please note that the problem you solved (isomorphic strings) is different from the reference solution provided (group anagrams). In the future, ensure you're solving the correct problem as specified in the assignment. 

For the isomorphic strings problem, your solution is excellent. Keep up the good work!

PR #2201, opened 2025-09-17
Solved Hashing-1 problems

aahiawasthee:


super30admin:
- Strengths:
  - Correct implementation of all three problems.
  - Good use of helper methods to modularize code.
  - Clear and descriptive variable names.
  - Correct time complexity analysis for all problems.
- Areas for Improvement:
  - The space complexity for `groupAnagrams` should be O(N*K), not O(1), as it stores all strings in the map.
  - Consider using a larger data type (e.g., `BigInteger` in Java) for the prime product in `groupAnagrams` to handle very large strings without overflow.
  - The prime array in `groupAnagrams` includes an extra prime (103) which is unnecessary since there are only 26 letters in the alphabet.

PR #2204, opened 2025-10-07
Hashing1

deept2604:


super30admin:
- Strengths:
  - Correct implementation of both problems.
  - Good use of prime multiplication for hashing in Group Anagrams.
  - Efficient use of a set to check for duplicate mappings in Isomorphic Strings.
  - Clear and readable code with appropriate comments.
- Areas for Improvement:
  - Remove commented-out code (Solution 1 in Group Anagrams) to improve readability.
  - Consider adding a brief docstring or comment explaining the prime multiplication approach in Group Anagrams for better clarity.
  - In Isomorphic Strings, the check `if len(s)==1` is redundant since the loop will handle it correctly without this special case.

PR #2205, opened 2025-10-07
Hashing-1

ashishbadhe97:


super30admin:
- Strengths:
  - The student has provided multiple solutions for Group Anagrams, demonstrating a good understanding of different approaches.
  - The student correctly analyzes the time and space complexity for each solution.
- Areas for Improvement:
  - For Group Anagrams, Solution 1 could be noted as less efficient due to sorting, while Solution 2 is more efficient and should be preferred.
  - In Isomorphic Strings, the use of both a HashMap and a HashSet is correct, but the student could consider using two HashMaps for symmetry, similar to the approach in Word Pattern.
  - The student could add more comments explaining the logic behind the key generation in Solution 2 of Group Anagrams (using the hashtable and converting it to a string).

PR #2216, opened 2025-10-13
Hashing-1 Submission

tanyasreenagesh:


super30admin:
Strengths:
- The solutions are correct and handle edge cases appropriately.
- The use of two maps for bijective checks in IsomorphicStrings.py and WordPattern.py is a good approach.

Areas for Improvement:
- For GroupAnagrams.py, consider using a hash function based on prime numbers (as in the reference solution) to avoid the sorting step and improve time complexity.
- While the space complexity is noted as O(1) for IsomorphicStrings.py and WordPattern.py, it might be more accurate to describe it as O(k) where k is the size of the character set, as the maps can grow up to the size of the character set.

PR #2227, opened 2025-11-18
Solved Hashing 1

nikhil2297:


super30admin:
Strengths:
- Good use of comments to explain the approach.
- Proper use of data structures (HashMap, ArrayList) in Problem3.java and Problem1.java.
- Efficient sorting and key generation in Problem1.java.

Areas for Improvement:
- Problem2.java needs to handle bidirectional mapping correctly. The current solution only checks one direction, which can lead to false positives. Consider using two maps like in the reference solution.
- The marked array in Problem2.java is not necessary if you use two maps, which would simplify the solution.
- While the time complexity is optimal in all cases, ensure the logic is correct before optimizing for performance.

PR #2233, opened 2025-11-20
Hashing1

Tejaswini2112:
word pattern, isomorphic strings and Grouping anagrams


super30admin:
Strengths:
- The student has provided both brute-force and optimized solutions, demonstrating a good understanding of the problem.
- The use of prime numbers for hashing is a clever and efficient approach.
- The student has handled edge cases and large numbers using BigInteger, which is a good practice.

Areas for Improvement:
- The brute-force approach, while correct, is less efficient and could be omitted if the focus is solely on the optimized solution.
- The student could consider adding more test cases to validate the solution further.
- The comments could be more concise while still being informative.

PR #2243, opened 2025-12-30
Completed Hashing-1

rbhargav0104:


super30admin:
Strengths:
- The solution correctly identifies and groups anagrams.
- The use of a hash map to group anagrams is a good approach.

Areas for Improvement:
- The time complexity can be improved by avoiding the sorting step. Consider using a hashing technique like the reference solution, which multiplies prime numbers corresponding to each character. This would reduce the time complexity from O(N * K log K) to O(N * K).
- While the current solution is correct, it may not be the most efficient for very large inputs due to the sorting overhead.

PR #2257, opened 2026-01-08
Group anagrams, isomorphic strings, pattern matching in strings

paridhimalviya:
1. Group anagrams
2. check if strings are isomorphic or not
3. pattern matching between a pattern string i.e. abba and the string "dog cat cat dog"

### DIFF
--- a/GroupAnagrams.swift
+++ b/GroupAnagrams.swift
@@ -1,0 +1,82 @@
+//
+//  GroupAnagrams.swift
+//  DSA-Practice
+//
+//  Created by Paridhi Malviya on 1/3/26.
+//
+
+//Same characters with same frequence but the order can be jumbeled
+class GroupAnagrams {
+    
+    init() {
+        
+        groupAnagrams()
+    }
+    
+    /*
+     In hash map, hashing is done for strings which are stored as keys. Under the hood, hashing can use multiple techniques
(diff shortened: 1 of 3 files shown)


super30admin:
Strengths:
- Good understanding of the problem and multiple approaches (sorting and prime product)
- Clear comments explaining the thought process
- Proper separation of concerns with helper methods

Areas for Improvement:
- The prime product implementation has a bug in the index calculation. It should be `asciiOfCh - asciiOfa` instead of the reverse.
- Method naming should be consistent (use camelCase for all method names)
- The prime product approach should be the main solution since it's more efficient
- Consider adding more test cases to verify edge cases
- The initialization and method calls could be better organized (currently the solutions are run in init methods)

super30admin:
Strengths:
1. The student has a good understanding of the problem and has attempted two different approaches.
2. The code is well-commented and structured, making it easy to follow.
3. The student correctly identifies the space complexity and discusses the trade-offs between different approaches.

Areas for Improvement:
1. The prime product method has a critical error in the index calculation. The line `let index = asciiOfa - asciiOfCh` should be `let index = asciiOfCh - asciiOfa` to correctly map characters to their respective primes.
2. The time complexity of the sorted strings method is higher than the optimal solution. The student should focus on optimizing this further.
3. The solution could benefit from more rigorous testing, especially edge cases like empty strings or strings with non-alphabetic characters.
4. The student should ensure that all methods are correctly implemented and tested before submission.

PR #2261, opened 2026-02-03
[IMP] Hashing-1

amitmittal117:


super30admin:
Your solution for grouping anagrams is correct and well-written. The approach of using sorted strings as keys in a dictionary is straightforward and effective. However, note that the time complexity of your solution is O(n * k log k), which is acceptable for the given constraints (with k up to 100). 

For comparison, the reference solution uses a prime number hashing technique to achieve O(n * k) time complexity. This method avoids the log k factor by using multiplication of prime numbers to represent each string. While your solution is correct, in an interview setting you might be asked about alternative methods, so it's good to be aware of the prime hashing approach.

One potential issue with the prime hashing method is that for very long strings, the product of primes might become very large and cause integer overflow. However, in Python, integers can be arbitrarily large, so that might not be a problem. Alternatively, you could use a tuple of character counts as the key, which would also be O(n * k) but with a higher constant factor.

Your code is clean and easy to understand. Well done!

For the other two problems (isomorphic strings and word pattern), your solutions are also correct and efficient. However, please note that the problem we are evaluating is specifically the anagrams problem.

PR #2266, opened 2026-02-10
Complete Hashing-1

PrakarshKamal:


super30admin:
Your solution for grouping anagrams is correct and efficient. You have correctly identified that anagrams have the same character frequency, and you used that to create a key for the hash map. The code is clean and well-commented.

One minor point: the key generation using `Arrays.toString(freq)` is correct, but note that it creates a string that includes brackets and commas, which is acceptable. However, be aware that if the frequency counts are large (e.g., above 9), the string might have more characters, but since the array length is fixed at 26, it is still constant. Alternatively, you could use a string builder to create a custom string representation (e.g., with delimiters) to avoid any remote chance of collision (though with integer frequencies, there is no collision). But your current method is safe.

Another common approach is to use a string of characters representing the frequency, but your method is straightforward and correct.

Overall, great job!

super30admin:
Your solution for GroupAnagrams is correct and efficient. It correctly groups anagrams by using a frequency array and a hash map. The time and space complexities are optimal. The code is clean and readable.

One minor suggestion: Instead of using `putIfAbsent` and then adding to the list, you can use `computeIfAbsent` to make the code more concise. For example:
```java
map.computeIfAbsent(key, k -> new ArrayList<>()).add(str);
```
This avoids the need for two operations (putIfAbsent and then get) and is more idiomatic.

Also, note that the key generated by `Arrays.toString(freq)` will be a string like "[1, 0, 0, ...]" which is unique for each frequency pattern. This is correct, but be aware that the string might be long (26 integers) but since it's fixed, it's acceptable.

Overall, great job! Your solution is correct and well-implemented.

PR #2274, opened 2026-02-20
Hashing 1 done

pranjay01:


super30admin:
Your solution for the Group Anagrams problem is excellent. It correctly groups anagrams by using a frequency array and a dictionary with tuple keys. The time and space complexity are optimal. The code is clean and well-commented.

One minor suggestion: remove the print statement at the end of the code. In a coding interview or when submitting to an online judge, the solution should only contain the class and method definitions without any external test calls. This ensures that the code runs without unexpected output.

Also, note that in Python, the ord('a') is 97, but you can also use ord('a') to make it more readable. However, using 97 is acceptable.

Keep up the good work!

PR #2275, opened 2026-02-24
Done Hashing -1

gurneetk186:


super30admin:
Your solution is well-implemented and follows a common approach for this problem. Here are some points to consider:
- The code is clear and concise. Good job on using the sorted string as the key.
- You might consider using the interface type `Map` instead of the concrete class `HashMap` for the declaration to adhere to best practices in Java (e.g., `Map<String, List<String>> map = new HashMap<>();`).
- For very large inputs, the sorting approach might be slightly slower than alternative methods (like using a frequency count string or prime number hashing), but for the given constraints, it is efficient.
- Note that the prime number hashing method (as in the reference solution) has a risk of integer overflow or floating point precision issues for very long strings, but with double and primes, it might work for strings up to 100 characters. Your sorting method avoids such risks.

Overall, your solution is correct and efficient.

PR #2276, opened 2026-02-26
Solved Hashing-1

pratul2789:


super30admin:
Your solution for Group Anagrams is well thought out and uses the prime number hashing technique correctly. The code is clean and readable. However, there are a few points to note:

1. In Python, it is better to return a list of the dictionary values explicitly. Instead of `return d.values()`, use `return list(d.values())` to ensure compatibility with Python 3 and to meet the problem's requirement of returning a list of lists.

2. Although not necessary for the given constraints, be aware that for very long strings the product of primes might become very large. However, in Python integers can handle this without overflow, so it is acceptable.

3. You might consider adding a comment about why you chose primes and how the hashing works, but the code is clear enough.

4. The solution for Isomorphic Strings was included in the same file? Actually, the student provided two separate files. But in the submission, it seems like they are both in the same response. For the purpose of this problem, we are only evaluating Group Anagrams. Make sure to submit only the required solution.

5. For the empty string, your code works correctly because the hash for an empty string is 1.

Overall, great job! Just remember to return a list explicitly.

PR #2291, opened 2026-05-04
PRs for Hashing 1 (Problem 1 and 2)

nikhylw:
Problem 1 - W2_2_49_group_anagrams.py
Problem 2 - W2_1_205_Isomorphic_Strings.py


super30admin:
### Grouping Anagrams Together  (W2_2_49_group_anagrams.py)
Your solution is well-structured and follows a good approach. However, there are a few points to consider:
1. **Hash Function**: Using a product of primes for hashing is a valid technique, but in Python, it's common to use a tuple of character counts or a sorted string as the key to avoid large integers. For example, you could use `key = tuple(sorted(s))` which is straightforward and avoids potential issues with large numbers. Alternatively, if you prefer the prime method, note that for long strings the product might become very large, which could impact performance in terms of memory and computation time. However, since the problem constraints state that the string length is at most 100, it might be acceptable, but it's good to be aware of this.

2. **Edge Cases**: Your solution should handle the empty string correctly. The prime product method returns 1 for an empty string, which is fine, but test with examples like `[""]` to ensure it works.

3. **Code Clarity**: Your code is clear and readable. You've included complexity comments, which is excellent. Consider adding a docstring for the `get_hash` method to explain what it does.

4. **Alternative Approach**: Another common approach is to use a frequency array (tuple of counts) for each string. For example:
   ```python
   def get_hash(s: str) -> tuple:
        count = [0] * 26
        for c in s:
            count[ord(c) - ord('a')] += 1
        return tuple(count)
   ```
   This method is also efficient and avoids large integers.

Overall, your solution is correct and efficient for the given constraints, but consider the alternative key methods for better practice.

VERDICT: PASS

---

### Isomorphic Strings (W2_1_205_Isomorphic_Strings.py)
Your solution is well-structured and correctly solves the problem. You have used appropriate data structures (a dictionary for mapping and a set to track used characters) to ensure the constraints are met. The time and space complexity are optimal.

Strengths:
- You correctly identified the need to check both that each character in s maps to exactly one character in t and that each character in t is mapped to by at most one character in s.
- Your use of `zip(s, t)` is Pythonic and efficient.
- The comments are helpful in explaining your approach.

Areas for improvement:
- While your solution is correct, you can consider using two dictionaries (one for s->t and one for t->s) to make the code more symmetric and avoid maintaining a separate set. This is the approach taken in the reference solution. It might be slightly more efficient in terms of memory (because a set is essentially a dictionary without values, so it uses similar space) but it is also more straightforward.
- There is a minor issue with your comment placement: the comment "# Execute when current character is already in the map" is placed above the `else` block, which is correct. However, the indentation in your submitted code might have been slightly off in the text representation. Always ensure that your code is properly indented when submitting.

Here is an alternative

VERDICT: NEEDS_IMPROVEMENT

---

### Word Pattern 
Dear student,

It seems you have submitted solutions for different problems (Isomorphic Strings and Group Anagrams) instead of the Word Pattern problem. Please note that each problem requires a specific approach.

For the Word Pattern problem, you need to:
1. Split the input string `s` into a list of words.
2. Check if the number of words equals the length of the pattern. If not, return false.
3. Use two dictionaries (or hash maps): 
   - One to map each character in the pattern to a word.
   - Another to map each word to a character in the pattern.
4. Iterate through each character in the pattern and each word in the list simultaneously.
5. For each pair (char, word):
   - If the char is already in the first dictionary, check that the mapped word matches the current word.
   - If the word is already in the second dictionary, check that the mapped char matches the current char.
   - If either condition fails, return false.
   - Otherwise, add the mappings to both dictionaries.

Here is a Python example for the correct solution:
```python
def wordPattern(pattern, s):
    words = s.split()
    if len(pattern) != len(words):
        return False
    
    char_to_word = {}
    word_to_char = {}
    
    for c, word in zip(pattern, words):
        if c in char_to_word:
            if char_to_word[c] != word:
                return False
        else:
            char_to_word[c] = word
            
        if word in word_to_char:
            if word_to_char[word] != c:
                return False
        else:
            word_to_char[word] = c
            
    return True
```

Please ensure you are solving the correct problem and review the problem statement carefully. Your solutions for the other problems are interesting, but they are not relevant here.

VERDICT: NEEDS_IMPROVEMENT

PR #2293, opened 2026-06-09
Solved Hashing-1 Assignment

vvkumaryadla:
Solved Hashing-1 Assignment


super30admin:
### Grouping Anagrams Together  (groupAnaGrams.java)
**Strengths:**
- Clear problem-solving approach with good explanation
- Correct algorithm implementation
- Good variable naming

**Areas for Improvement:**
1. Remove the redundant `map.put(sortedString, value)` line after modifying the list - Java's HashMap stores references, so the update is already reflected
2. Use `computeIfAbsent` for cleaner code:
   ```java
   map.computeIfAbsent(sortedString, k -> new ArrayList<>()).add(str);
   ```
3. Simplify by returning `new ArrayList<>(map.values())` directly instead of iterating
4. Consider the trade-off: sorting approach is simpler to understand, but the prime-hashing approach in the reference is more efficient for strings with many characters

VERDICT: PASS

---

### Isomorphic Strings (IsIsomorphic.java)
**Strengths:**
- Good defensive programming with the length check
- Correctly implements the isomorphic string logic

**Areas for Improvement:**
1. **Time Complexity**: The `map.values().contains(tArray[i])` operation is O(n) per iteration, making the overall complexity O(n²). Consider using a second HashMap for reverse mapping (like the reference solution) to achieve O(n) time complexity.
2. **Alternative Approach**: You could also sort the characters and compare patterns, or use index tracking, but the HashMap approach with reverse mapping is the most efficient.

**Suggested Optimization:**
```java
HashMap<Character, Character> reverseMap = new HashMap<>();
// In the loop, after mapping s->t, also check/update reverseMap
```

This would eliminate the O(n) lookup for checking existing values.

VERDICT: NEEDS_IMPROVEMENT

---

**Strengths:**
1. Clean, readable implementation with good variable names
2. Correct algorithm logic that properly handles the bijection constraint
3. Good use of HashMap for mapping storage
4. Clear inline comments explaining the approach

**Areas for Improvement:**
1. **Time Complexity**: The solution is O(n²) due to the `patternMap.values().contains(words[i])` check which scans all values. Using a second HashMap (like the reference solution) would make this O(1), reducing overall complexity to O(n).
2. **Space Complexity**: The comment mentions O(n·m) which is unnecessarily complex. It should simply be O(n) where n is the number of unique pattern characters.
3. **Documentation**: The time complexity comment contains a typo ("n^2.m" should be "n²·m") and is incorrect. Consider updating to reflect actual complexity.

**Optimization Suggestion:**
Replace the `patternMap.values().contains(words[i])` check with a second HashMap for reverse mapping. This changes the lookup from O(n) to O(1), improving overall efficiency.

VERDICT: PASS

